PR #5, opened 2017-08-02
Expose function to parse URI from parsed certificates

csstaub:
Proposed change to make this library easier to consume. I'd like to hook this up in [square/certigo](https://github.com/square/certigo) so that certigo can display URI SANs when dumping a certificate, but I don't want to round-trip to the PEM-encoded version. There's also some `gofmt` changes (related: you should apply `gofmt` automatically on all commits).

Note this change slightly alters the API due to the fact that I renamed `GetUrisInSubjectAltName` to `GetUrisInSubjectAltNameEncoded`, let me know if that's not ok and I can change the naming to something else. 

@amartinezfayo @y2bishop2y @evan2645

### DIFF
--- a/spiffe.go
+++ b/spiffe.go
@@ -2,9 +2,9 @@ package spiffe
 
 import (
 	"crypto/x509"
-	"encoding/pem"
-	"encoding/asn1"
 	"crypto/x509/pkix"
+	"encoding/asn1"
+	"encoding/pem"
 	"errors"
 )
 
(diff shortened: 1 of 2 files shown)


csstaub:
The link for the CLA is broken for me (404 - Not Found). Where can I find it?

evan2645:
This is awesome! Thank you @csstaub!!

I think we're working on getting the CLA fixed up today

@amartinezfayo is there any precedence for the function names we're using here? We might benefit from some easier-to-grok naming (as long as we're changing the API, and while we still can!)

amartinezfayo:
Thanks @csstaub for this. I think that this change makes a lot of sense. Regarding the naming, I just tried to choose a descriptive name. We can certainly explore better alternatives.

evan2645:
Proposed: `GetURIsFromCertificate` and `GetURIsFromPEM`... thoughts?

amartinezfayo:
@evan2645 do you think that this will be specific enough? My concern is that this may be ambiguos with getting for example URIs from the name constraints extension vs getting from SAN.

evan2645:
@amartinezfayo that's a fair point... though ideally the user won't deal with name constraint URIs and we'd lean on chain validation to address that concern.

My objection to the current naming is that it exposes the "guts" of the certificate. In particular, a SPIFFE user is unlikely to know anything about Subject Alt Name. Maybe then `GetURINamesFromCertificate` better addresses the concern? Or perhaps I am being pedantic :)

amartinezfayo:
Got it @evan2645. Then I think that GetURIsFromCertificate or GetURINamesFromCertificate are good candidates. I would personally vote for GetURINamesFromCertificate.

evan2645:
@csstaub any preference from you?

csstaub:
`GetURINamesFromCertificate` sounds good to me.

evan2645:
I cherry picked this commit in #8 

Thanks again for the contrib! Please let us know how we can help getting this up in https://github.com/square/certigo

PR #11, opened 2017-08-11
all: added CLI, FGetURINamesFromPEM and godoc reference

### LINKED ISSUE #7: How can I `go get` and use this library?
For convenience, suggest publishing guidance on how to go-get this library, and publish a code snippet detailing how it was used.

odeke-em:
Fixes #7

- Added copy+pastable examples to show the package can be used, in
example_test.go.

- Added a godoc reference so that users can get the latest
references of the code without us having to manually update
the README.md file each time.

- Implemented a CLI that is go gettable and can be used on the
commandline to inspect URIs in certificates for example
```shell
$ go get github.com/spiffe/go-spiffe/cmd/spiffe
$ spiffe $HOME/certs/leaf.cert.pem
Path:: #1: "testdata/leaf.cert.pem"
  URI #1: "spiffe://dev.acme.com/path/service"
```

- Implemented a convenience method that takes in an io.Reader which
can then be easily used in programs that say read from files, as the
CLI does
```go
package main

import (
  "io"
  "log"
  "os"

  spiffe "github.com/spiffe/go-spiffe"
)

func main() {
  log.SetFlags(0)

  var f io.Reader
  if len(os.Args) <= 1 {
    // Expecting to read from stdin e.g
    // $ cat
    f = os.Stdin
  } else {
    rf, err := os.Open(os.Args[1])
    if err != nil {
      log.Fatal(err)
    }
    defer rf.Close()
    f = rf
  }

  uris, err := spiffe.FuriNamesFromPEM(f)
  if err != nil {
    log.Fatal(err)
  }
  if len(uris) == 0 {
    log.Fatal("did not get any URIs")
  }
  for i, uri := range uris {
    log.Printf("#%d:: URI:%q", i, uri)
  }
}
```

which when run gives
```shell
$ cat $GOPATH/src/github.com/spiffe/go-spiffe/testdata/leaf.cert.pem | go run spiffe-cli.go
\#0:: URI:"spiffe://dev.acme.com/path/service"
```

### DIFF
--- a/cmd/spiffe/main.go
+++ b/cmd/spiffe/main.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"log"
+	"os"
+
+	spiffe "github.com/spiffe/go-spiffe"
+)
+
+func main() {
+	// Make log not print out time information
+	log.SetFlags(0)
+
+	if len(os.Args) <= 1 {
+		log.Fatalf("\033[31mexpecting path [paths...] e.g\033[00m\n\tspiffe /var/lib/letsencrypt/spiffe.pem $HOME/Desktop/certs/cert.pem\033[00m\n")
+	}
+
+	seenPaths := make(map[string]bool)
+	args := os.Args[1:]
+	for i, path := range args {
+		if _, seen := seenPaths[path]; seen {
+			continue
+		}
+
+		seenPaths[path] = true
+		f, err := os.Open(path)
+		if err != nil {
+			log.Printf("\n#%d: err=%v\n", i, err)
+			continue
+		}
+
+		uris, err := spiffe.FGetURINamesFromPEM(f)
+		_ = f.Close()
+
+		if err != nil {
+			log.Printf("\n#%d: parseErr=%v\n", i, err)
+			continue
+		}
+		if len(uris) == 0 {
+			log.Printf("\n#%d: no uris could be parsed\n", i)
+			continue
+		}
+
+		log.Printf("Path:: #%d: %q\n", i+1, path)
+		for j, uri := range uris {
+			log.Printf("\tURI #%d: %q\n", j+1, uri)
+		}
+	}
+}
--- a/spiffe.go
+++ b/spiffe.go
@@ -6,6 +6,8 @@ import (
 	"encoding/asn1"
 	"encoding/pem"
 	"errors"
+	"io"
+	"io/ioutil"
 )
 
 var oidExtensionSubjectAltName = asn1.ObjectIdentifier{2, 5, 29, 17}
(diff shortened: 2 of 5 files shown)


odeke-em:
Exhibits:

### Before
screen shot 2017-08-10 at 11 34 57 pm

### After
screen shot 2017-08-10 at 11 33 32 pm

/cc @ajessup who created that issue.

odeke-em:
So I signed the CLA sent to my corp email, about 1 hour ago, but it hasn't yet updated here, PTAL!

cmd/spiffe/main.go
```diff
@@ -0,0 +1,49 @@
...
+
+func main() {
+	// Make log not print out time information
+	log.SetFlags(0)
+
+	if len(os.Args) <= 1 {
+		log.Fatalf("\033[31mexpecting path [paths...] e.g\033[00m\n\tspiffe /var/lib/letsencrypt/spiffe.pem $HOME/Desktop/certs/cert.pem\033[00m\n")
+	}
+
+	seenPaths := make(map[string]bool)
```
evan2645: It seems that we can simplify this a bit by taking only one arg instead of many. I'm curious if you have a use for examining multiple certs in one invocation?

spiffe.go
```diff
@@ -93,3 +101,12 @@ func GetURINamesFromPEM(encodedCertificate string) (uris []string, err error) {
 
 	return GetURINamesFromCertificate(cert)
 }
+
+// FuriNamesFromPEM retrieve URIs from the SAN extension of the reader containing a PEM-encoded X.509 certificate
+func FuriNamesFromPEM(f io.Reader) (uris []string, err error) {
```
evan2645: Can we update this function name to align with the rest of the exported functions? Perhaps `FGetURINamesFromPEM`?

drrt:
@odeke-em not to worry - the bot only does its checks when the PR is submitted or updated with new commits. You've been added to the contributors config, and we'll just ignore the check when the PR is merged.

cmd/spiffe/main.go
odeke-em: Good question @evan2645, am examining multiple certs because that creates a nice use case for the CLI, that way when testing, one can supply multiple certs on the CLI which is easier to do than writing a unit test.

spiffe.go
odeke-em: Interestingly I actually had it before as `FgetURINamesFromPEM` but thought the `get` was redundant, however I'll update that. Thanks for letting me know.

odeke-em:
@drrt, awesome, thanks for letting me know. I'll update the PR with @evan2645's requested change and then move on from there.

odeke-em:
@evan2645, PTAL, I've updated the PR with the requested function name change.

odeke-em:
Thanks y'all for the review. For sure, I look forward to making more in the future!

PR #12, opened 2017-08-28
Add functions that can help to parse X.509 SVID certificates

amartinezfayo:
- Add MarshalUriSANs function. This can help to construct a CSR with a URI SAN.
- Add GetURINamesFromExtensions function. This can help to parse the URI SAN from a CSR.
- Add GetKeyUsageExtensionsFromCertificate function. This can help to validate X.509 SVID certificates.
- Expose OID identifiers
- Fixed some function descriptions.

### DIFF
--- a/spiffe.go
+++ b/spiffe.go
@@ -10,7 +10,8 @@ import (
 	"io/ioutil"
 )
 
-var oidExtensionSubjectAltName = asn1.ObjectIdentifier{2, 5, 29, 17}
+var OidExtensionSubjectAltName = asn1.ObjectIdentifier{2, 5, 29, 17}
+var OidExtensionKeyUsage = asn1.ObjectIdentifier{2, 5, 29, 15}
 
 func getURINamesFromSANExtension(sanExtension []byte) (uris []string, err error) {
 	// RFC 5280, 4.2.1.6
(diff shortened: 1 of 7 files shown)


spiffe.go
```diff
@@ -111,3 +112,33 @@ func FGetURINamesFromPEM(f io.Reader) (uris []string, err error) {
...
 	return uriNamesFromPEM(blob)
 }
+
+// GetURINamesFromExtensions retrieves URIs from the SAN extension of a slice of extensions
+func GetURINamesFromExtensions(extensions *[]pkix.Extension) (uris []string, err error) {
+	for _, ext := range *extensions {
+		if ext.Id.Equal(OidExtensionSubjectAltName) {
+			uris, err = getURINamesFromSANExtension(ext.Value)
+			if err != nil {
+				return
```
evan2645: Can we be explicit w/ the return vals here? We are in most other occasions, and I think it's easier to read that way anyways

amartinezfayo: :+1: I will change it.

amartinezfayo:
Yes, I think it makes sense to divide the packages now. I'll create a package `uri` for all the uri handling. The `spiffe` package will remain for the rest, to handle spiffe-specific stuff. The sooner we make this change, the better.

PR #15, opened 2017-09-05
Add tls helpers

evan2645:
This PR:
* Adds TLS helpers for callback and mTLS config
* Moves OID-specific stuff to dedicated file in uri package
* Adds spiffe-specific helpers to validate chains and match spiffe ids

The exact validation methodology for SPIFFE is still in flux, so leaving a TODO there where we can hook it in once we're ready


evan2645:
PS>> Part of the reasoning for moving the OID stuff around (apart from it not really being spiffe-specific) is that the current organization can result in a circular dependency

evan2645:
Good call @amartinezfayo - fixed

PR #24, opened 2019-09-21
Fix comments on spiffe id validation modes

jkanywhere:
Add or remove path to make sure every example given for a
spiffe.ValidationMode is a valid SPIFFE ID according to that mode.

### DIFF
--- a/spiffe/id.go
+++ b/spiffe/id.go
@@ -140,7 +140,7 @@ func AllowAny() ValidationMode {
 	return validationMode{}
 }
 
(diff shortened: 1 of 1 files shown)


jkanywhere:
You're welcome. For anyone else who, like me, is not familiar with DCO, it means Developer Certificate" and refers to the text "Signed-off-by: Name <email>" in the commit message.
You can learn about this custom here: https://blog.chef.io/2016/09/19/introducing-developer-certificate-of-origin/ and https://developercertificate.org/

PR #29, opened 2019-11-28
Panic prevention and test coverage

jkanywhere:
Prevent panic when validating or normalizing a nil *url.URL as SPIFFE ID.

While in the package, add tests to increase coverage of spiffe/id.go.


spiffe/id_test.go
```diff
@@ -160,19 +162,99 @@ func TestValidateID(t *testing.T) {
...
+			}
 			assert.NoError(t, err)
+
 			assert.Equal(t, test.out, out)
 		})
 	}
 }
+
+func TestNormalizeURI(t *testing.T) {
+	t.Run("nil", func(t *testing.T) {
```
jkanywhere: This test would fail on master.

spiffe/id.go
```diff
@@ -48,8 +48,12 @@ func ValidateURI(id *url.URL, mode ValidationMode) error {
...
+		var idStr string
+		if id != nil {
+			idStr = id.String()
+		}
 		return fmt.Errorf("invalid %sSPIFFE ID %q: "+format,
-			append([]interface{}{kind, id.String()}, args...)...)
+			append([]interface{}{kind, idStr}, args...)...)
 	}
 
 	if id == nil || *id == (url.URL{}) {
```
jkanywhere: The `id == nil` case was not previously covered.

PR #74, opened 2020-03-13
Implement X.509 bundle type

marcosy:
This PR implements the X.509 bundle type of the v2 API

Fixes: #55 

Signed-off-by: Marcos G. Yedro <marcosyedro@gmail.com>


v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,182 @@
...
 // AddX509Root adds an X.509 root to the bundle. If the root already
 // exists in the bundle, the contents of the bundle will remain unchanged.
-func (b *Bundle) AddX509Root(*x509.Certificate) {
-	panic("not implemented")
+func (b *Bundle) AddX509Root(root *x509.Certificate) {
+	b.rootsMtx.Lock()
+	defer b.rootsMtx.Unlock()
+
+	for _, r := range b.roots {
+		if reflect.DeepEqual(r, root) {
```
azdagron: Reflect is probably overkill here. Also, we compare certificates in a few spots... probably good to make a helper function out of it for readability.
```suggestion
		if bytes.Equal(r.Raw, root.Raw) {
```

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,182 @@
...
+
+	if b.trustDomain != trustDomain {
+		return nil, errors.New("wrong td")
+	}
+
+	return b, nil
+}
+
+// removeElement removes an element from slice, slice order is not preserved.
+func removeElement(slice []*x509.Certificate, index int) ([]*x509.Certificate, error) {
```
azdagron: This seems to be only used in one spot, and that spot has a valid index.... I'd recommend either moving the body into RemoveX509Root or getting rid of the bounds checking since it is statically valid.

v2/bundle/x509bundle/bundle_test.go
```diff
@@ -0,0 +1,216 @@
...
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_New(t *testing.T) {
```
azdagron: Conventional naming guidelines don't have an underscore in the `Test` method names (i.e. `TestNew` instead of `Test_New`). The exception are examples to distinguish between different variants (e.g. `ExampleNew` and `ExampleNew_someVariant`)

v2/bundle/x509bundle/bundle_test.go
```diff
@@ -0,0 +1,216 @@
...
+func Test_GetX509BundleForTrustDomainFails(t *testing.T) {
+	bundle, err := Load("example.org", "testdata/certs.pem")
+	require.NoError(t, err)
+
+	b, err := bundle.GetX509BundleForTrustDomain("another-td.org")
+	require.Error(t, err)
+	require.Nil(t, b)
+}
+
+func Test_removeElementSucceeds(t *testing.T) {
```
azdagron: Unless there is a really compelling reason, I'd rather not unit test internals, especially when they are testable via the public interface.

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,182 @@
...
+	"sync"
 
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
 )
 
+const certType string = "CERTIFICATE"
+
 // Bundle is a collection of trusted public key material for a trust domain.
 type Bundle struct {
+	tdMtx       *sync.RWMutex
```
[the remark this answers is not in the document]
marcosy: That's right! will fix

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,182 @@
...
+		rootsMtx:    &sync.RWMutex{},
+	}
 }
 
 // Load loads a Bundle from a file on disk.
 func Load(trustDomain spiffeid.TrustDomain, path string) (*Bundle, error) {
-	panic("not implemented")
+	fileBytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open file: %v", err)
```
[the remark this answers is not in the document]
amartinezfayo: I'm not sure if we should say unable to `open` when we are calling `ReadFile`. Maybe saying `unable to read` is more accurate? And also include the file path?
something like `fmt.Errorf("unable to read X.509 bundle from %q: %w", path, err)`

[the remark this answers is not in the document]
azdagron: The error message coming back from ReadFile already contains the path, e.g. `"open blah: no such file or directory"`.

v2/bundle/x509bundle/bundle.go
marcosy: By putting this in a separate function, my intention was to have better readability. But then I  thought that this method may be used in other places in the future and got concerned about index checking.

I believe I rather move the body into RemoveX509Root instead of getting rid of the bounds checking. Just in case.

v2/bundle/x509bundle/bundle_test.go
marcosy: Right! I copied from a wrong template 🤦‍♂

v2/bundle/x509bundle/bundle_test.go
marcosy: Got it. This was because there was no way to test "index out of bound" scenario using the public interface. I'll remove it.

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,182 @@
...
 // GetX509BundleForTrustDomain returns the X.509 bundle for the given trust
 // domain. It implements the Source interface. It will fail if
 // called with a trust domain other than the one the bundle belongs to.
 func (b *Bundle) GetX509BundleForTrustDomain(trustDomain spiffeid.TrustDomain) (*Bundle, error) {
-	panic("not implemented")
+	b.tdMtx.RLock()
+	defer b.tdMtx.RUnlock()
+
+	if b.trustDomain != trustDomain {
+		return nil, errors.New("wrong td")
```
amartinezfayo: we may be more specific here? saying "trust domain" instead of "td" and maybe include the wrong trust domain?

v2/bundle/x509bundle/bundle_test.go
```diff
@@ -0,0 +1,216 @@
...
+	bundle, err := Load("example.org", "testdata/non-existent-file.pem")
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "unable to open file")
+	assert.Nil(t, bundle)
+}
+
+func Test_ReadSucceeds(t *testing.T) {
+	file, err := os.Open("testdata/certs.pem")
+	require.NoError(t, err)
+	defer file.Close()
```
amartinezfayo: @azdagron I'm curious to know if this would be a place where the use of the newly supported cleanup functions is recommended? https://golang.org/pkg/testing/#T.Cleanup

v2/bundle/x509bundle/bundle.go
marcosy: Totally agree! I forgot to refine this error msg after development.

v2/bundle/x509bundle/bundle_test.go
azdagron: If this was a helper function that opened the file for testing, then Cleanup would be a good way of not making the caller have to arrange for cleanup. However, considering this is the test function itself, the traditional cleanup pattern using defer is just fine. Either way is appropriate.

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,182 @@
...
 )
 
+const certType string = "CERTIFICATE"
+
 // Bundle is a collection of trusted public key material for a trust domain.
 type Bundle struct {
+	tdMtx       *sync.RWMutex
+	trustDomain spiffeid.TrustDomain
+
+	rootsMtx *sync.RWMutex
```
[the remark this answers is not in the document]
marcosy: > nit: the prevailing guidance is that mutex fields in structs that are used via references (i.e. *Bundle) be values, not pointers.

Thanks! will fix

> I'm trying to image a scenario where a bundle consumer is also mutating the bundle...

Hmm... I've been thinking about his... One case that comes to mind is someone serving a bundle endpoint (not backed by the workload API), that runs a separate thread to update the roots certificates in the bundle source. Does it makes sense?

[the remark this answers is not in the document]
azdagron: Yeah, I think you are right. Considering this can be used as a bundle source and provided to a TLS config that will be hitting this thing from another goroutine, it's probably better to be safe.

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,175 @@
...
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
 )
 
+const certType string = "CERTIFICATE"
+
 // Bundle is a collection of trusted public key material for a trust domain.
 type Bundle struct {
+	trustDomain spiffeid.TrustDomain
+	roots       []*x509.Certificate
+	rootsMtx    sync.RWMutex
```
azdagron: nit: group fields protected by the mutex (with the mutex at the top)

```suggestion
	trustDomain spiffeid.TrustDomain
	
	rootsMtx    sync.RWMutex
	roots       []*x509.Certificate

```

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,175 @@
...
+		roots:       []*x509.Certificate{},
+	}
 }
 
 // Load loads a Bundle from a file on disk.
 func Load(trustDomain spiffeid.TrustDomain, path string) (*Bundle, error) {
-	panic("not implemented")
+	fileBytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to load X.509 bundle file: %w", err)
```
azdagron: See the review on #76. I think we want to prefix these errors with the package name (i.e. `x509bundle: `)

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,175 @@
...
+	}
+
+	return Parse(trustDomain, fileBytes)
 }
 
 // Read decodes a bundle from a reader.
 func Read(trustDomain spiffeid.TrustDomain, r io.Reader) (*Bundle, error) {
-	panic("not implemented")
+	var b bytes.Buffer
```
azdagron: nit: `ioutil.ReadAll` is probably a better fit here since we don't really need a buffer (just the bytes)

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,176 @@
...
+	rootsMtx sync.RWMutex
+	roots    []*x509.Certificate
 }
 
 // New creates a new bundle
 func New(trustDomain spiffeid.TrustDomain) *Bundle {
-	panic("not implemented")
+	return &Bundle{
+		trustDomain: trustDomain,
+		roots:       []*x509.Certificate{},
```
azdagron: nit: this doesn't seem necessary? `nil` should behave the same as an empty slice at least in the ways we're using it.... am I missing something?

v2/bundle/x509bundle/bundle.go
```diff
@@ -1,65 +1,176 @@
...
 // RemoveX509Root removes an X.509 root to the bundle.
-func (b *Bundle) RemoveX509Root(*x509.Certificate) {
-	panic("not implemented")
+func (b *Bundle) RemoveX509Root(root *x509.Certificate) {
+	b.rootsMtx.Lock()
+	defer b.rootsMtx.Unlock()
+
+	for i, r := range b.roots {
+		if areCertsEqual(r, root) {
+			//remove element from slice (slice order is not preserved)
```
azdagron: I've looked over this a few times and keep rationalizing how not preserving order should be ok, but every time it gives me pause. Considering removing a root should be an infrequent, likely non-bulk operation, how would you feel doing a less performant "remove" just to preserve order? Somebody might be surprised by the change in ordering if they are diff'ing serialized versions.

marcosy: That is a good point. Removing elements in that way will cause more headaches than benefits. I'll change it

v2/bundle/x509bundle/bundle.go
marcosy: You are right! I thought it was necessary but it is not

PR #134, opened 2020-05-04
Solve tls config flaky test

MarcosDY:
Prevent connection to be closed before test finished on tls config handshake tests

Signed-off-by: Marcos Yacob <marcos@scytale.io>


v2/spiffetls/tlsconfig/config_test.go
```diff
@@ -688,14 +688,19 @@ func testConnection(t testing.TB, serverConfig *tls.Config, clientConfig *tls.Co
 		}
 		require.NoError(t, err)
 	}()
+
+	complete := make(chan struct{})
```
amartinezfayo, line 692: nit: `done `instead of `complete`?

PR #159, opened 2021-02-05
add rpc to fetch x509 bundles

adobley:
This is the first step on the journey to complete
[#2089](https://github.com/spiffe/spire/issues/2089)

SPIRE will need to implement the handler for this rpc in
order for the client library to start using this call.

SPIRE relies on this repo for the definition and codegen so this needs
to be first.

Signed-off-by: Samuel Waggoner <swaggoner@vmware.com>
Co-authored-by: Samuel Waggoner <swaggoner@vmware.com>


azdagron (approved):
Looking good! Thanks @adobley !

Issue #174, opened 2021-11-23
Possible bug in getTLSCertificate where err might not be returned


yangmarcyang:
In v2/spiffetls/tlsconfig/config.go `getTLSCertificate` function, the `return nil, err` is put inside the `if trace.GotCertificate != nil` block. So, it sounds like a bug when trace.GotCertificate is not configured, where the error would not be returned and the following code would cause panic when dereferencing svid `s`.

Code pointer: [link](https://github.com/spiffe/go-spiffe/blob/7b2184ee742dec93be292cdf8cc06ac6fbffcdbd/v2/spiffetls/tlsconfig/config.go#L217)

If it is a real bug, the fix would look very simple: just move the return statement outside below the if block.

azdagron:
That is a bug!

azdagron:
You're suggested fix is spot on.

yangmarcyang:
> Your suggested fix is spot on.

Thanks for the confirmation. I will create a PR then.

azdagron:
Fixed by #175

PR #175, opened 2021-11-23
Fix no error return in getTLSCertificate

yangmarcyang:
Fix no error return in getTLSCertificate when trace function is not configured

Change-Id: I1ba680638e61006f48651c0c3b34c5a06e4e9399

Fixes issue #174


azdagron (approved):
Thanks!

PR #225, opened 2023-05-26
Export PeerIDFromConnectionState

sorindumitru:
This was mentioned on Slack that it could useful to have to be able extract the SPIFFE ID from a [http.Request](https://pkg.go.dev/net/http#Request) strucutre.


azdagron (approved):
Thanks, this has been a long time coming :) When the library was first written, we didn't want to export until we had a chance to maybe determine a better way. But that effort never bore fruit and I myself have copied this half a dozen times into various projects.

PR #232, opened 2023-06-22
ci(deps): setup dependabot

mmorel-35:
Setup dependabot to keep gomod and github-actions dependencies up-to-date. 
This will open pull requests for each update that you can review.
My pr includes the config of dependabot to check for updates once a week.
It also includes the update of all available dependencies. 

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


bleggett (approved):
This LGTM and would solve my issue around outdated deps in this package causing a downstream `go get` downgrade loop.

azdagron (approved):
Thanks @mmorel-35 !

PR #255, opened 2023-12-19
Export fake workload and bundle APIs

keeganwitt:
Exporting the fake workload and bundle APIs will allow them to be used in issues like spiffe/spiffe-helper#112.


keeganwitt:
Should we also export [v2/internal/test/ca.go](https://github.com/spiffe/go-spiffe/blob/main/v2/internal/test/ca.go) (and [v2/internal/test/keys.go](https://github.com/spiffe/go-spiffe/blob/main/v2/internal/test/keys.go))?

keeganwitt:
Notes from contributors sync meeting:

It was felt it might be risky to export this, since it might be inadvertently changed and unclear that there are other projects depending on the implementation, and to instead rely on https://github.com/spiffe/spire/tree/main/test/fakes/fakeworkloadapi (though that doesn't support the bundle API yet). There is also an internal tool that might be open sourced as a separate library for this functionality.

PR #270, opened 2024-03-06
Enable race detection when running tests

strideynet:
Adds `-race` to the invocation of `go test` in the `test` target of the Makefile. This is generally considered a good practice as it enables catching potential races earlier.

Looks like the test suite runs happily with the race detector enabled 🥳 

Whilst this can reduce the speed at which the test suite runs, it doesn't look like this produces a significant change for this test suite on my local machine, especially given the relatively small size of the suite at the moment.


azdagron (approved):
Thanks @strideynet !

PR #276, opened 2024-03-30
Update to go-jose v4.0.1

rturner3:
There are some breaking changes in v4 that affect go-spiffe code:
- When parsing a JWS-protected JWT, the caller must now explicitly pass the accepted signature algorithms (corresponding to `alg` JWS header) into the parsing function.
- `CompactSerialize()` was removed in favor of just `Serialize()`, which has identical behavior.

go-jose v4 updates the minimum Go version requirement to 1.21, which was released 2023-08-08.


v2/go.mod
```diff
@@ -1,10 +1,12 @@
 module github.com/spiffe/go-spiffe/v2
 
-go 1.19
+go 1.21
+
+toolchain go1.22.1
```
rturner3, line 5: This was auto-added by `go mod tidy` when run with Go v1.22.1. I can remove it if it is going to cause any backward compatibility issues and still compatible with latest tooling.

rturner3:
This PR depends on #277 to be merged first.

azdagron (approved):

PR #277, opened 2024-04-01
Bump Go to 1.21

rturner3:
The latest version of `github.com/go-jose/go-jose` depends on Go 1.21+. Since Go 1.21 has been released since March 2023, it should be safer to use it for building/testing code in the repo.

Also bump `golangci-lint` to latest version 1.57.2, as the older version used in the repo is incompatible with Go 1.21+. Some additional changes to make the lint build pass:

- Exclude `depcheck` linter that was previously enabled, but not configured. It now causes the lint build to fail since no dependencies are configured in the allowlist.
- Disable `unused-parameter` rule in `revive` linter, since naming parameters can be helpful for documentation purposes in libraries.
- Reorganize `golangci-lint` config that was moved in newer versions.
- Fix usage of rand source in Windows `fakeworkloadapi` that was depending on a deprecated `rand.Seed()` method as well as a weak random number generator from `math/rand`.

### DIFF
--- a/.github/workflows/pr_build.yaml
+++ b/.github/workflows/pr_build.yaml
@@ -3,7 +3,7 @@ on:
   pull_request: {}
   workflow_dispatch: {}
 env:
-  GO_VERSION: 1.19
+  GO_VERSION: 1.21
 jobs:
   lint-linux:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


azdagron (approved):
Thanks, @rturner3 !

PR #357, opened 2025-09-06
Make it possible to mock the current time in spiffetls dialer

vandry:
So you can do, say,

  spiffetls.DialWithMode(
      ctx, network, addr, mode,
      spiffetls.WithDialTLSOptions(tlsconfig.WithTime(mocked))
  )

This is useful in tests and is already supported by the underlying crypto/x509 library and even by the x509svid package right here but just wasn't plumbed through to the dialer.


azdagron:
Closing stale PR.

Issue #365, opened 2025-10-09
WIT-SVID: Add a witbundle package


amartinezfayo:
Similarly to the jwtbundle and x509bundle packages, we need to add bundle support for the WIT-SVID profile.

jsnctl:
Hi @amartinezfayo, I've put together a patch for what I understand the changes to `go-spiffe` might be here but I wanted to check my thinking before I open a PR against the repo https://github.com/spiffe/go-spiffe/compare/main...jsnctl:jsnctl/wit-bundle

The bundle implementation herein is mostly a lift-and-shift and s/jwt/wit for `jwtbundle` -> `witbundle`. I think this is correct, as while obviously the WIT-SVID will be divergent from JWT-SVID, the format of the authorities in the bundle will be the same(?)

I've wrapped the `jwtutils` implementation in the case of `witutils`, and this could be done throughout if we wanted to avoid the duplication and greatly reduce the size of this diff.

amartinezfayo:
Hey @jsnctl thank you for working on this!

>the format of the authorities in the bundle will be the same(?)

Yes, that's correct. Both JWT-SVID and WIT use JWKS (RFC 7517) format for their signing keys. The WIT spec uses standard JOSE headers and the bundle stores the verification keys in the same format as JWT bundles.

>I've wrapped the jwtutils implementation in the case of witutils, and this could be done throughout if we wanted to avoid the duplication and greatly reduce the size of this diff.

I think that's the right approach.

The patch is looking good. I think that you can open a draft PR and we can finish the review once https://github.com/spiffe/spiffe/pull/327 is merged.

arndt-s:
Resolved through #385

PR #367, opened 2025-10-16
Add WIT bundle functionality

jsnctl:
implements https://github.com/spiffe/go-spiffe/issues/365

Opening draft PR ahead of merge of upstream https://github.com/spiffe/spiffe/pull/327 (per discussion [here](https://github.com/spiffe/go-spiffe/issues/365#issuecomment-3412263337))


exp/witbundle/bundle.go
```diff
@@ -0,0 +1,202 @@
...
+	defer b.mtx.RUnlock()
+
+	_, ok := b.witAuthorities[keyID]
+	return ok
+}
+
+// AddWITAuthority adds a WIT authority to the bundle. If a WIT authority already exists
+// under the given key ID, it is replaced. A key ID must be specified.
+func (b *Bundle) AddWITAuthority(keyID string, witAuthority crypto.PublicKey) error {
+	if keyID == "" {
```
strideynet, line 116: The constraint that a keypair must have a `kid` is interesting - I have raised this for discussion on https://github.com/spiffe/spiffe/pull/327#discussion_r2439044279

Today, the JWT-SVID/WIT-SVID profiles of the Workload API do not make the `kid` member of the JWKS mandatory - but it looks like `go-spiffe` treats it as such. I do think that the behaviour of `go-spiffe` here is reasonable, but, we should consider at least modifying the WIP WIT-SVID profile to express that the `kid` member is mandatory.

exp/witbundle/bundle.go
```diff
@@ -0,0 +1,202 @@
+package witbundle
```
arndt-s, line 1: @jsnctl I wonder if we should keep everything WIT-SVID related in an `exp` folder to indicate clearly that this is still experimental? Otherwise there's no differentiation between X509, JWT and WIT SVID and users of this SDK get the perception that this is fully supported.

Once the RFC is finished and has a number we can then move it into the main part of the SDK.

exp/witbundle/bundle.go
```diff
@@ -0,0 +1,202 @@
...
+func FromWITAuthorities(trustDomain spiffeid.TrustDomain, witAuthorities map[string]crypto.PublicKey) *Bundle {
+	return &Bundle{
+		trustDomain:    trustDomain,
+		witAuthorities: witutil.CopyWITAuthorities(witAuthorities),
+	}
+}
+
+// Load loads a bundle from a file on disk. The file must contain a standard RFC 7517 JWKS document.
+func Load(trustDomain spiffeid.TrustDomain, path string) (*Bundle, error) {
+	bundleBytes, err := os.ReadFile(path)
```
arndt-s, line 43: nit: you could use `Read()` here and supply the file as io.Reader.

exp/witbundle/set.go
arndt-s, line 1: This should ideally be a generic. Curious to get the maintainers POV on this. But maybe something for another day.

exp/witbundle/bundle.go
jsnctl: Looking at the existing `jwtsvid` implementation, it also concretely relies on the kid in the `FindJWTAuthority` call for verification (the corresponding `witsvid` changes will be in a separate merge)

https://github.com/spiffe/go-spiffe/blob/c03307b73fb29459cd4c150fbbf7d092995ad10b/svid/jwtsvid/svid.go#L53-L69

Presumably JWKS usage without a kid in each authority would just require iterating over all available JWKs, so this seems like a nice pattern to maintain here.

As for the spec, you can imagine (less tidy) downstream implementations which use valid JWKS without the kid? Perhaps the opinion of `go-spiffe` here isn't the only valid consideration?

strideynet: > Perhaps the opinion of go-spiffe here isn't the only valid consideration?

Perhaps so - although it definitely carries a fair bit of weight in terms of the "real-world" usage of the specification. Today, if a Workload API implementation does not provide the `kid` member of the trust bundle JWKS, then it looks like `go-spiffe`/`jwtbundle` would fail to unmarshal it. Fortunately, I think all "real-world" implementations of the SPIFFE specification are likely to include the `kid` member.

However, as we're introducing a new profile for WIT, it does seem like a good opportunity to make the spec a little more restrictive if it means that all proper implementations of the spec will be compatible with the SDKs. At the moment the gray area where an implementation could be spec-compliant but not work with `go-spiffe` seems a little problematic.

arndt-s (commented):
I started a review last week but got distracted. Publishing my comments so far.

exp/witbundle/bundle.go
jsnctl: Yeah good plan. I'm on holiday at the moment but I'll amend once I'm back

jsnctl: For explicitness: do you think this should be lifted to `bundle/exp/witbundle`, or should we add an `exp` folder at the root of the project which would also be used by the other parts of the WIT implementation?

arndt-s: I believe the latter is better and more consistent. Then all wit-related stuff is coherent in the `exp` folder.

strideynet: Agreed - we should probably add a statement to the README as well at somepoint (or a GoDoc in the `exp` package) to indicate that the usual guarantees of compatibility do not apply

jsnctl: I've added the new `exp` package, and added a brief `doc.go` to it https://github.com/jsnctl/go-spiffe/blob/jsnctl/wit-bundle/exp/doc.go

exp/witbundle/bundle.go
jsnctl: I agree this would be a nicer implementation - do we want to keep things consistent with the methods in the other `*bundle` implementations though?

https://github.com/spiffe/go-spiffe/blob/c03307b73fb29459cd4c150fbbf7d092995ad10b/bundle/jwtbundle/bundle.go#L41-L49

https://github.com/spiffe/go-spiffe/blob/c03307b73fb29459cd4c150fbbf7d092995ad10b/bundle/x509bundle/bundle.go#L40-L47

https://github.com/spiffe/go-spiffe/blob/c03307b73fb29459cd4c150fbbf7d092995ad10b/bundle/spiffebundle/bundle.go#L55-L64

strideynet: I think consistency is likely more important here - that being said - it'd be nice if we did extend the API of all the profiles to include `LoadFromReader` or similar? and perhaps refactor `Load` to leverage that under the hood.

jsnctl:
We'd originally said to hold off a merge until https://github.com/spiffe/spiffe/pull/327 was merged (which it is now), do with think that's still applicable here?

jsnctl:
Not sure of the best way to stack this, but I've also got a branch based off this diff for the corresponding `witsvid` changes too https://github.com/jsnctl/go-spiffe/pull/1

amartinezfayo:
Closing as superseded: #385 added the experimental `exp/bundle/witbundle` package, and #391 covers the remaining APIs from this PR that #385 did not include. Thank you @jsnctl for the contribution!

PR #379, opened 2026-03-26
Fix current build break

azdagron:
Updates the linter and Go download URLs and versions. Stays with go1.24 for now just to unblock CI.


amartinezfayo (commented):
Hey @azdagron, sorry I missed this. Looking at the linter output in the build, there are some new gosec rules that need to be fixed. I noticed #378 addresses those gosec issues so we may pick those from there.

amartinezfayo:
@azdagron did you have a chance to look at this?

PR #385, opened 2026-04-20
Add WIT-SVID

arndt-s:
This is the go-spiffe branch I've used for the KubeCon demo of WIT-SVIDs.


workloadapi/client.go
```diff
@@ -252,6 +256,263 @@ func (c *Client) ValidateJWTSVID(ctx context.Context, token, audience string) (*
...
+			continue
+		}
+		hints[s.Hint] = struct{}{}
+
+		svid, err := witsvid.ParseInsecure(s.WitSvid)
+		if err != nil {
+			return nil, err
+		}
+
+		if s.WitSvidKey != "" {
```
amartinezfayo: I'm wondering if the empty WitSvidKey case in parseWITSVIDs should error rather than produce an SVID with no PrivateKey. I think the proto declares the field as Required, and the WIT-SVID profile says that all WITSVID fields except `hint` are mandatory, and a WIT-SVID without the matching private key cannot be used for proof of possession (I believe that's the whole point of the credential).
A server returning one without the key looks more like a server bug worth surfacing than something that can be ignored. What do you think about returning an error when WitSvidKey is empty here?

workloadapi/wit_client_test.go
```diff
@@ -0,0 +1,307 @@
...
+
+	select {
+	case bundleSet := <-updateCh:
+		require.True(t, bundleSet.Has(witTD))
+	case <-time.After(5 * time.Second):
+		t.Fatal("timed out waiting for initial WIT bundle update")
+	}
+}
+
+func TestWatchWITSVIDs_Unimplemented(t *testing.T) {
```
amartinezfayo: I think TestWatchWITSVIDs_Unimplemented may not actually exercise the codes.Unimplemented path. The fake fetchWITSVID returns noIdentityError (which is PermissionDenied) when no response is configured, and handleWatchError retries on PermissionDenied, so what the test ends up asserting is that the 200ms context timeout fires.
To exercise the early-exit behavior, would it make sense to teach the fake to return codes.Unimplemented from FetchWITSVID/FetchWITBundles and assert that the watch returns promptly with that status code?

exp/bundle/witbundle/bundle.go
```diff
@@ -0,0 +1,131 @@
...
+func New(trustDomain spiffeid.TrustDomain) *Bundle {
+	return &Bundle{
+		trustDomain:    trustDomain,
+		jwtAuthorities: make(map[string]crypto.PublicKey),
+	}
+}
+
+// FromJWTAuthorities creates a new bundle from a map of JWT authorities keyed
+// by key ID.
+func FromJWTAuthorities(trustDomain spiffeid.TrustDomain, jwtAuthorities map[string]crypto.PublicKey) *Bundle {
```
amartinezfayo: I wonder if the JWTAuthorities/FindJWTAuthority/AddJWTAuthority naming on witbundle.Bundle ends up being a bit confusing once spiffebundle.Bundle is in the picture, since the latter exposes the WIT-flavored counterparts as WITAuthorities/FindWITAuthority/etc. The mismatch is most visible in spiffebundle.FromWITBundle, which calls witBundle.JWTAuthorities() to populate bundle.witAuthorities, and in witsvid.ParseAndValidate, which calls bundle.FindJWTAuthority(keyID) on a witbundle to verify a WIT-SVID.
Would naming the witbundle methods WITAuthorities/FindWITAuthority, etc. make the two APIs feel more consistent?

workloadapi/client.go
```diff
@@ -252,6 +256,263 @@ func (c *Client) ValidateJWTSVID(ctx context.Context, token, audience string) (*
...
+// Experimental: subject to change.
+func (c *Client) FetchWITSVIDs(ctx context.Context, spiffeID string) ([]*witsvid.SVID, error) {
+	return c.fetchWITSVIDs(ctx, spiffeID, false)
+}
+
+// WatchWITSVIDs watches for WIT-SVID updates. The watcher receives the updated
+// WIT-SVIDs. The optional spiffeID filters updates to a specific identity.
+//
+// If the server returns codes.Unimplemented, the watch loop terminates without
+// retrying per the WIMSE spec §7.
```
amartinezfayo: The Unimplemented-then-no-retry behavior is defined in the SPIFFE Workload API spec under §7 "WIT-SVID Profile", whereas the WIMSE documents specify the WIT token format itself rather than the workload API protocol.
Would it make sense to retitle the "WIMSE spec §7" citation here (and on WatchWITBundles)?
Pointing at "SPIFFE Workload API §7 (WIT-SVID Profile)" instead would attribute it correctly I think.

exp/svid/witsvid/svid.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil, wrapErr(fmt.Errorf("unable to parse claims: %v", err))
+	}
+
+	publicKey, err := extractCnfJWK(rawClaims)
+	if err != nil {
+		return nil, err
+	}
+
+	// Verify expiry
+	if stdClaims.Expiry.Time().Before(time.Now()) {
```
amartinezfayo: Maybe we could also enforce the `nbf` claim in parse.
From what I understand in the WIT-SVID spec 3.5, it says `nbf` MAY be present and that "validators MUST reject WIT-SVIDs when the time indicated by `nbf` is in the future". The current parse only enforces `exp`, so a token whose `nbf` is in the future would be accepted.
Would extracting `nbf` alongside `exp` from `stdClaims` and rejecting when it lies after `time.Now()` work here?

exp/svid/witsvid/svid.go
```diff
@@ -0,0 +1,211 @@
...
+	if !ok {
+		return nil, wrapErr(errors.New("cnf claim missing jwk field"))
+	}
+
+	jwkMap, ok := jwkRaw.(map[string]interface{})
+	if !ok {
+		return nil, wrapErr(errors.New("cnf.jwk is not an object"))
+	}
+
+	if _, ok := jwkMap["alg"]; !ok {
```
amartinezfayo: What do you think about extending the check here from a presence test to a value check?
I think the spec lists the specific `alg` values the `cnf.jwk.alg` MUST take (the same RS/ES/PS suite already in `allowedSignatureAlgorithms` for the JOSE header). Without that enforcement, a token whose confirmation key declares an unsupported algorithm would parse here and only fail later when the workload tries to use it for proof of possession.

exp/svid/witsvid/svid.go
```diff
@@ -0,0 +1,211 @@
...
+// (e.g., the Workload API).
+func ParseInsecure(token string) (*SVID, error) {
+	return parse(token, func(*jwt.JSONWebToken, spiffeid.TrustDomain, string) error {
+		return nil
+	})
+}
+
+type verifyFn func(*jwt.JSONWebToken, spiffeid.TrustDomain, string) error
+
+func parse(token string, verify verifyFn) (*SVID, error) {
```
amartinezfayo, line 98: It could be worth adding direct _test.go files for the new witsvid and witbundle packages. It seems that the parse failure modes in witsvid.parse are only reached transitively through wit_client_test.go, and the witbundle helpers aren't directly covered at all.
The test coverage in svid/jwtsvid could be a useful template I think.

workloadapi/client.go
```diff
@@ -252,6 +256,263 @@ func (c *Client) ValidateJWTSVID(ctx context.Context, token, audience string) (*
...
+		bundles = append(bundles, b)
+	}
+
+	return witbundle.NewSet(bundles...), nil
+}
+
+func parseJWKPrivateKey(jwkStr string) (interface{}, error) {
+	var jwkKey jose.JSONWebKey
+	if err := jwkKey.UnmarshalJSON([]byte(jwkStr)); err != nil {
+		return nil, fmt.Errorf("witsvid: unable to parse private key JWK: %v", err)
```
amartinezfayo: nit: the "witsvid:" prefix in these errors looks a bit off. The function lives in the workloadapi package, and the rest of the WIT-related parsing in this file emits errors without that prefix, so the current message reads as if it came from a different layer than it actually does.

amartinezfayo (commented):
This is great @arndt-s! Thank you so much for this contribution.
I have some comments/suggestions.

exp/svid/witsvid/svid.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, wrapErr(fmt.Errorf("unable to parse claims: %v", err))
+	}
+
+	publicKey, err := extractCnfJWK(rawClaims)
+	if err != nil {
+		return nil, err
+	}
+
+	// Verify expiry
+	if stdClaims.Expiry.Time().Before(time.Now()) {
```
amartinezfayo: It seems that `exp` and `nbf` checks here don't apply clock-skew leeway, whereas the `jwtsvid` package validates through go-jose's `Claims.Validate`, which applies a one-minute `DefaultLeeway` to both `exp` and `nbf`. The WIT-SVID spec says validators MAY allow a small amount of leeway for clock skew, and  frames `nbf` similarly.
As the logic reads, a WIT-SVID would be rejected for skew that a JWT-SVID would tolerate, which seems like an inconsistency worth aligning. Would it make sense to apply a small leeway (or reuse `jwt.Expected`/`ValidateWithLeeway`) so the two SVID types behave the same under skew?

workloadapi/witsource.go
```diff
@@ -0,0 +1,249 @@
...
+	}()
+
+	if err := s.waitForInitial(ctx); err != nil {
+		return nil, err
+	}
+	s.drainUpdated()
+	return s, nil
+}
+
+func (s *WITSource) waitForInitial(ctx context.Context) error {
```
amartinezfayo, line 79: nit: Could `WITSource` reuse the shared `watcher` in watcher.go instead of hand-rolling its lifecycle? The cancel func, `wg`, per-stream `sync.Once` set-channels, `updatedCh`, `drainUpdated`, and `triggerUpdated` seem very similar to what `watcher` already provides, so the duplication means a future fix has to land in two places.
I realize the shared watcher is currently shaped around X509Context and JWTBundles, so generalizing it may be out of scope for this PR.

amartinezfayo (commented):
Looking great! Thank you @MarcosDY for addressing previous comments.
I've added a couple of suggestions.

amartinezfayo (approved):
Thank you @arndt-s and @MarcosDY! 🚀

exp/svid/witsvid/svid.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+// ParseAndValidate parses and validates a WIT-SVID token, verifying its
+// signature using the provided WIT bundle source.
+func ParseAndValidate(token string, bundles witbundle.Source) (*SVID, error) {
+	return parse(token, func(tok *jwt.JSONWebToken, trustDomain spiffeid.TrustDomain, keyID string) error {
+		bundle, err := bundles.GetWITBundleForTrustDomain(trustDomain)
+		if err != nil {
+			return wrapErr(fmt.Errorf("no WIT bundle found for trust domain %q", trustDomain))
+		}
```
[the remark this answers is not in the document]
MarcosDY: intentionally we are avoiding to return errors

workloadapi/option.go
```diff
@@ -49,6 +49,20 @@ type SourceOption interface {
 	configureX509Source(*x509SourceConfig)
 	configureJWTSource(*jwtSourceConfig)
 	configureBundleSource(*bundleSourceConfig)
+	configureWITSource(*witSourceConfig)
+}
```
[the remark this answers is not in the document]
MarcosDY: internal use, no affecting

PR #388, opened 2026-05-15
Add Broker API

matheuscscp:
Supersedes: https://github.com/spiffe/spire-api-sdk/pull/92 https://github.com/spiffe/spire-api-sdk/pull/96

SPIFFE PRs: https://github.com/spiffe/spiffe/pull/340 https://github.com/arndt-s/spiffe/pull/1
SPIRE PR: https://github.com/spiffe/spire/pull/6915

Hey @arndt-s, @sorindumitru suggested adding this directly here instead of `spire-api-sdk` since this comes directly from SPIFFE and not SPIRE.


exp/proto/spiffe/broker/api.proto
strideynet, line 1: Need to discuss with `go-spiffe` maintainers, but we had discussed in SIG-SPEC generating this within an `exp/` directory until the spec was stable.

matheuscscp: Are these the maintainers? https://github.com/spiffe/go-spiffe/blob/main/CODEOWNERS

@amartinezfayo @evan2645 @azdagron @MarcosDY

matheuscscp: Should I simply move all new files in this PR under `exp/` in the root dir?

matheuscscp: @strideynet Guidance pls :pray:

strideynet: Sadly, I can't provide the answer you are looking for - since I don't "maintain" the `go-spiffe` project - we'd need one of the maintainers to chime in with their preferences as to how we handle code that implements experimental/draft specifications.

matheuscscp: @strideynet Thanks, can you pls confirm the maintainers are the people in CODEOWNERS?

strideynet: Yep, you've tagged the right people. Augustin and Marcos are the most active though - but you may need to give them a few days.

amartinezfayo: In #367 we agreed to place experimental functionality under a root-level exp/ folder. I think the same approach fits here: moving the new proto package under exp/proto/spiffe/broker/ (and exp/proto/spiffe/reference/ as well), plus a short doc.go at the exp/ root calling out that the usual compatibility guarantees do not apply.

I'd suggest updating the go_package options to match the new paths while keeping the proto package declarations unchanged, so the wire identity does not churn when we promote them out of exp/.

matheuscscp: Fixed, thanks!

matheuscscp:
The spec PR just got merged, this one is ready to go!

exp/proto/spiffe/broker/api.proto
```diff
@@ -0,0 +1,215 @@
...
+
+// The SubscribeToX509SVIDResponse message carries X.509-SVIDs and related information,
+// including a list of bundles the workload may use for federating with foreign
+// trust domains.
+message SubscribeToX509SVIDResponse {
+    // Required. A list of X509SVID messages, each of which includes a single
+    // X.509-SVID, its private key, and the bundle for the trust domain.
+    repeated X509SVID svids = 1;
+
+    // Optional. ASN.1 DER encoded certificate revocation lists.
```
amartinezfayo: I think the `brokerapi.proto` that was just merged includes a `(not PEM)` qualifier on every DER-encoded field comment to make it unambiguous for implementors, but that appears to be missing from `crl` here and throughout the file (`federated_bundles` in `SubscribeToX509SVIDResponse`; `x509_svid`, `x509_svid_key`, and `bundle` in `X509SVID`; and both `crl` and `bundles` in `SubscribeToX509BundlesResponse`).
I think it would be good to carry over the exact comment text from the spec.

exp/proto/spiffe/broker/api.proto
```diff
@@ -0,0 +1,215 @@
...
+// a vendor-specific extension type.
+message WorkloadReference {
+    // Required. The reference to the workload.
+    google.protobuf.Any reference = 1;
+}
+
+// The WorkloadPIDReference message conveys a process id reference of a workload
+// running in the same environment.
+message WorkloadPIDReference {
+    // Required. The process id of the workload. MUST be a positive integer.
```
amartinezfayo, line 49: The `brokerapi.proto` in the spec appends a second sentence to this comment that addresses the container-sandbox case: "For workloads running inside container runtimes that maintain a separate sandbox (e.g. the `pause` container in containerd or CRI-O), this MUST be the PID of the workload's own process, not the sandbox PID." I think that guidance would be worth carrying here too since it covers a common implementation pitfall for Kubernetes-based deployments.

exp/doc.go
```diff
@@ -0,0 +1,6 @@
+// Package exp contains experimental implementations of go-spiffe components
+// * proto/spiffe/broker - SPIFFE Broker API protobuf definitions
+//
+// Note: Code herein is may be subject to breaking changes, and should not be
```
amartinezfayo: It looks like there's a typo here: "Code herein is may be subject to breaking changes" should probably read "Code herein may be subject to breaking changes".

exp/proto/spiffe/broker/api.proto
matheuscscp: Fixed, thanks!

exp/proto/spiffe/broker/api.proto
matheuscscp: Fixed, thanks!

exp/doc.go
matheuscscp: Fixed, thanks!

amartinezfayo (approved):
🎉

PR #393, opened 2026-06-05
spiffeid: enforce 2048-byte maximum SPIFFE ID length

### LINKED ISSUE #384: spiffeid: SPIFFE ID max length (2048 bytes) not enforced
## Summary

The `spiffeid` package does not enforce the 2048-byte maximum length limit required by the SPIFFE specification:

> The total length of a SPIFFE ID MUST NOT exceed 2048 bytes.
> — [SPIFFE-ID spec §2](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md#2-spiffe-identity)

## Affected functions

All ID constructors are affected since they share the same unchecked helpers:

- `FromPath` / `FromPathf` / `FromSegments`
- `FromString` / `FromStringf` / `FromURI`
- `AppendPath` / `AppendPathf` / `AppendSegments`
- `ReplacePath` / `ReplacePathf` / `ReplaceSegments`
- `TrustDomainFromString` / `TrustDomainFromURI`

## Root cause

`makeID` assembles the final SPIFFE ID string but performs no length check.
`ValidatePath` validates character set and segment structure only.
`TrustDomainFromString` similarly has no length guard.

## Expected behavior

Any function that produces or parses a SPIFFE ID should return an error when
the resulting ID exceeds 2048 bytes.

## Suggested fix

Add a length check in `makeID` (the single convergence point for all constructors):

```go
const maxIDLength = 2048

func makeID(td TrustDomain, path string) (ID, error) {
    if td.IsZero() {
        return ID{}, errors.New("trust domain is empty")
    }
    id := schemePrefix + td.name + path
    if len(id) > maxIDLength {
        return ID{}, errIDTooLong
    }
    return ID{
        id:      id,
        pathidx: schemePrefixLen + len(td.name),
    }, nil
}
```
FromString parses without going through makeID, so it would need a separate length check as well.

arpitjain099:
Fixes #384

## Problem

The `spiffeid` package never enforces the SPIFFE specification's 2048-byte maximum length for a SPIFFE ID. An ID well over the limit can be constructed and parsed without error, so a library consumer can produce IDs that other SPIFFE implementations are not required to accept.

`FromPath(td, "/" + strings.Repeat("a", 3000))` and the equivalent `FromString(...)` both return a 3021-byte ID with a nil error today.

## Root cause

`makeID` is the convergence point for most constructors but has no length check. `FromString` parses directly and bypasses `makeID`, and the `Append*` methods mutate `id.id` directly and also bypass it.

## Fix

Added `const maxIDLen = 2048` and an `errIDTooLong` sentinel, with a length check in:

- `makeID` (covers `FromPath`, `FromPathf`, `FromSegments`, `ReplacePath`/`f`, `ReplaceSegments`, `TrustDomain.ID()`)
- `FromString` (covers `FromStringf`, `FromURI`, and the trust-domain URI path transitively)
- `AppendPath`, `AppendPathf`, `AppendSegments`

Trust-domain constructors do not need a separate guard (a bare trust domain is at most a few hundred bytes once wrapped; the RFC 3986 255-byte trust-domain limit is a distinct concern and out of scope for this issue). Valid IDs (<= 2048 bytes) are unaffected.

## Tests

Added `TestIDMaxLength` with explicit boundaries (exactly 2048 accepted, 2049 rejected) across `FromPath`, `FromString`, `FromSegments`, `AppendPath`, `AppendPathf`, and `AppendSegments`, plus a normal-ID sanity check. Before the change the over-length construction succeeds; after it returns the error. `go test ./spiffeid/...` and the whole-module `go test ./...` pass; `go vet` and gofmt are clean.

## A note on spec wording

The SPIFFE ID spec says a SPIFFE Server "SHOULD NOT generate URIs of length greater than 2048 bytes" and that implementations "MUST support URIs up to 2048 bytes." Returning a hard error on construction is therefore slightly stricter than the literal "SHOULD NOT," and it prevents this library from generating IDs that are not guaranteed to interoperate. If you would prefer to keep this non-fatal (for example gated behind an option, or a warning), I am happy to adjust; I went with a hard error since #384 asks for the limit to be enforced.

### DIFF
--- a/spiffeid/id.go
+++ b/spiffeid/id.go
@@ -10,6 +10,12 @@ import (
 const (
 	schemePrefix    = "spiffe://"
 	schemePrefixLen = len(schemePrefix)
+
+	// maxIDLen is the maximum length, in bytes, of a SPIFFE ID. The SPIFFE
+	// specification recommends that implementations not generate SPIFFE IDs
+	// longer than 2048 bytes so that they remain interoperable.
+	// See https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md#23-maximum-spiffe-id-length
+	maxIDLen = 2048
 )
 
 // FromPath returns a new SPIFFE ID in the given trust domain and with the
@@ -73,6 +79,10 @@ func FromString(id string) (ID, error) {
 
 	if err := ValidatePath(id[pathidx:]); err != nil {
 		return ID{}, err
+	}
+
+	if len(id) > maxIDLen {
+		return ID{}, errIDTooLong
 	}
 
 	return ID{
@@ -154,6 +164,9 @@ func (id ID) AppendPath(path string) (ID, error) {
 		return ID{}, err
 	}
 	id.id += path
+	if len(id.id) > maxIDLen {
+		return ID{}, errIDTooLong
+	}
 	return id, nil
 }
 
@@ -170,6 +183,9 @@ func (id ID) AppendPathf(format string, args ...interface{}) (ID, error) {
 		return ID{}, err
 	}
 	id.id += path
+	if len(id.id) > maxIDLen {
+		return ID{}, errIDTooLong
+	}
 	return id, nil
 }
 
@@ -186,6 +202,9 @@ func (id ID) AppendSegments(segments ...string) (ID, error) {
 		return ID{}, err
 	}
 	id.id += path
+	if len(id.id) > maxIDLen {
+		return ID{}, errIDTooLong
+	}
 	return id, nil
 }
 
@@ -251,8 +270,12 @@ func makeID(td TrustDomain, path string) (ID, error) {
 	if td.IsZero() {
 		return ID{}, errors.New("trust domain is empty")
 	}
+	id := schemePrefix + td.name + path
+	if len(id) > maxIDLen {
+		return ID{}, errIDTooLong
+	}
 	return ID{
-		id:      schemePrefix + td.name + path,
+		id:      id,
 		pathidx: schemePrefixLen + len(td.name),
 	}, nil
 }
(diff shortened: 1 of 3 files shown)


spiffeid/id.go
```diff
@@ -75,6 +81,10 @@ func FromString(id string) (ID, error) {
 		return ID{}, err
 	}
 
+	if len(id) > maxIDLen {
```
amartinezfayo: I think we should be careful here and scope the enforcement to generation only, dropping this guard in FromString. FromString is our acceptance path: it backs FromURI in x509svid/verify.go and the subject parse in jwtsvid/svid.go, so a length check here becomes a verification rule and would make us reject a peer's cryptographically-valid, properly-signed SVID solely because its ID is long. The spec frames 2048 as "SHOULD NOT generate" plus "MUST support up to 2048 bytes," with no requirement to reject longer IDs on parse, so rejecting them trades interoperability for no security gain and runs against being liberal in what we accept.
I think that the generation-side checks in makeID and the Append* methods cover what #384 is really after, since those are paths where local code constructs IDs it controls and erroring early can't cause an interop regression. If we ever do want the verify side to reject over-length IDs, I'd prefer we make that its own deliberate, documented behavior change rather than fold it into this one.
What do you think?

amartinezfayo (commented):
Thanks for the contribution, @arpitjain099!

arpitjain099:
Thanks for the careful review, @amartinezfayo, that makes sense. I've scoped the check to the generation paths and dropped it from FromString, so it no longer acts as a verification rule and we won't reject a peer's valid, signed SVID just because the ID is long. makeID and the Append* methods keep the guard, which covers what #384 was after without any interop regression on parse. Tests pass (go test ./... green, vet and gofmt clean on the package); happy to adjust further if you'd like.

PR #395, opened 2026-06-17
Add wit-svid to marshaled WIT bundle keys

MarcosDY:
The SPIFFE WIT spec (§6.1/§6.2) requires that each JWK entry in a WIT bundle JWKS document carry "use": "wit-svid".
 This field is how consuming clients distinguish WIT authorities from other keys that may coexist in a shared JWKS document. Without it, a client has no reliable way to identify which keys are valid WIT authorities.

### DIFF
--- a/exp/bundle/witbundle/bundle.go
+++ b/exp/bundle/witbundle/bundle.go
@@ -16,6 +16,9 @@ import (
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
 )
 
+// JWKUse is the JWK use value for WIT authorities in a JWKS document.
+const JWKUse = "wit-svid"
+
 // Bundle is a collection of trusted WIT authorities for a trust domain.
 type Bundle struct {
 	trustDomain spiffeid.TrustDomain
@@ -165,6 +168,7 @@ func (b *Bundle) Marshal() ([]byte, error) {
 		jwks.Keys = append(jwks.Keys, jose.JSONWebKey{
 			Key:   witAuthority,
 			KeyID: keyID,
+			Use:   JWKUse,
 		})
 	}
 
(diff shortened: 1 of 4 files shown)


exp/bundle/witbundle/bundle.go
```diff
@@ -165,6 +168,7 @@ func (b *Bundle) Marshal() ([]byte, error) {
 		jwks.Keys = append(jwks.Keys, jose.JSONWebKey{
```
amartinezfayo, line 167: It could be worth pinning down the empty-bundle output here. Because Keys starts nil, an empty bundle marshals to {"keys":null}, and spiffebundle.Parse rejects that with "no authorities found" since it treats a nil keys member as absent, whereas {"keys":[]} parses as an empty set. Initializing Keys to a non-nil empty slice would let an empty bundle (for instance a revocation) round-trip through spiffebundle. I notice jwtbundle.Marshal has the same nil-Keys pattern, so if we change it here it may be worth aligning both for consistency.

exp/bundle/witbundle/bundle.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
 )
 
+// JWKUse is the JWK use value for WIT authorities in a JWKS document.
+const JWKUse = "wit-svid"
```
amartinezfayo: It might be worth unexporting this, since JWKUse is referenced only by the test and the sibling packages either keep their use values unexported (spiffebundle) or omit them entirely (x509bundle, jwtbundle). Asserting the literal `"wit-svid"` in the test would also guard the value §6.1 mandates, whereas asserting against the constant is self-referential since Marshal sets Use from the same constant. Would an unexported witSVIDUse plus the literal in the test work, unless you are anticipating external callers that build a WIT JWKS by hand?

amartinezfayo (approved):
🎉

PR #400, opened 2026-07-13
Added SPIFFE Bundle Map

ValFadeev:
This implements SPIFFE Bundle Map, following the [standard](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md#5-spiffe-bundle-map) definition.


bundle/spiffebundle/bundlemap.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return ParseBundleMap(b)
+}
+
+// ParseBundleMap parses a bundle map from bytes. The data must be a document
+// following the SPIFFE Bundle Map format.
+func ParseBundleMap(bundleMapBytes []byte) (*BundleMap, error) {
+	doc := &bundleMapDoc{}
+	if err := json.Unmarshal(bundleMapBytes, doc); err != nil {
```
bcleenders, line 63: This feature would be helpful for me as well! Thanks for suggesting to add it!

I did have one note; the current implementation skips the constraint:

> Consumers MUST reject maps in which duplicate trust domain name keys are detected, provided that the underlying libraries are capable of detecting such duplications.

The Go JSON unmarshalling does not reject such maps. Maybe that is sufficiently accepted with the second half of that sentence, but if the flagship implementation doesn't respect that rule then the line in the spec would seem a bit pointless.

There's also a slightly more difficult way to get a collision; trust domains can be of URL form (`spiffe://my-domain.com`) or domain name (`my-domain.com`). The `TrustDomainFromString` method (called in this file on line 75) would treat these as the same:
https://github.com/spiffe/go-spiffe/blob/e9973f6314a3fa0e36eb1f00fbfe37bdc1554b96/spiffeid/trustdomain.go#L14-L18

ValFadeev: Thank you for pointing this out, it was an oversight in the implementation! To your second point, I proposed to compare on parsed trust domain name. This is because the intention is not to have unique strings as map keys, but rather prevent ambiguity over which key material belongs to which trust domain. Hope this makes sense.

PR #405, opened 2026-07-23
Reject leaf certificates with root-path SPIFFE IDs in Verify

bcleenders:
`Verify` (and by extension `ParseAndVerify`) accepted leaf certificates whose SPIFFE ID had no path component (e.g. `spiffe://example.org`). X509-SVID spec section 5.2 requires that the leaf SPIFFE ID has a non-root path:

> The validator [...] MUST ensure that the SPIFFE ID has a non-root path component.

This check already existed in `validateLeafCertificate`, which is used by `Parse`/`ParseRaw`/`Load`, but `Verify` had its own inline leaf checks that were an incomplete copy. The duplication also meant `Verify` was missing the `digitalSignature` key usage requirement.

To avoid future inconsistencies between the two paths, this PR uses the same `validateLeafCertificate` function as the `Parse` path uses.

Related: https://github.com/spiffe/go-spiffe/pull/375

Spec reference: https://github.com/spiffe/spiffe/blob/main/standards/X509-SVID.md#52-leaf-validation


bcleenders:
I don't see test failures in CI, but locally I get:

```console
$ go test ./...
[...]
--- FAIL: TestVerify (0.00s)
    --- FAIL: TestVerify/bad_leaf_no_non-root_path_SPIFFE_ID (0.00s)
        verify_test.go:141:
            	Error Trace:	/Users/bcleenders/src/github/spiffe/go-spiffe/svid/x509svid/verify_test.go:141
            	Error:      	An error is expected but got nil.
            	Test:       	TestVerify/bad_leaf_no_non-root_path_SPIFFE_ID
FAIL
FAIL	github.com/spiffe/go-spiffe/v2/svid/x509svid	0.553s
```

svid/x509svid/verify_test.go
```diff
@@ -86,31 +94,31 @@ func TestVerify(t *testing.T) {
 			name:   "no URI SAN",
 			chain:  leaf1NoURI,
 			bundle: bundle1,
-			err:    "x509svid: could not get leaf SPIFFE ID: certificate contains no URI SAN",
+			err:    "x509svid: cannot get leaf certificate SPIFFE ID: certificate contains no URI SAN",
```
bcleenders, line 97: Updating the error messages, to align with `validateLeafCertificate`.

bcleenders:
@amartinezfayo could you take a look?

bcleenders:
@MarcosDY can I bother you for a review?

strideynet:
@amartinezfayo - I wonder if we need to bundle a bunch of these spec-correction (but breaking changes) together into one release? There's probably quite a few other breaking changes we need to make.

strideynet (commented):
From a code perspective, this change looks good - thank you for the contribution.

I think we need to have a discussion about how to safely ship these kinds of changes though (which is why I've not jumped straight into approval) - there's quite a few outstanding PRs that could be considered "breaking" changes and we probably want to coordinate to ship these together in a single release.

bcleenders:
Thanks for the update! That makes sense. Let me know if I can help!

PR #406, opened 2026-07-23
Add support for ID prefix authorization

### LINKED ISSUE #7151: spire broker api identity
The broker api uses the spire-agent's identity when serving the broker api. Since this svid has to be unique to a node, it can be hard to validate when connecting to it, especially when load balanced in a tcp config I think.

matheuscscp:
Closes: https://github.com/spiffe/spire/issues/7151

### DIFF
--- a/spiffeid/match.go
+++ b/spiffeid/match.go
@@ -1,6 +1,9 @@
 package spiffeid
 
-import "fmt"
+import (
+	"fmt"
+	"strings"
+)
 
 // Matcher is used to match a SPIFFE ID.
 type Matcher func(ID) error
@@ -19,6 +22,19 @@ func MatchID(expected ID) Matcher {
 			return fmt.Errorf("unexpected ID %q", actual)
 		}
 		return nil
+	})
+}
+
+// MatchIDPrefix matches any SPIFFE ID with the given ID prefix. A matching ID
+// must be in the same trust domain and have either the same path as the prefix
+// or a path with the prefix on a segment boundary. If the prefix has no path,
+// any ID in the same trust domain matches.
+func MatchIDPrefix(expected ID) Matcher {
+	return Matcher(func(actual ID) error {
+		if actual.MemberOf(expected.TrustDomain()) && matchPathPrefix(actual.Path(), expected.Path()) {
+			return nil
+		}
+		return fmt.Errorf("unexpected ID %q", actual)
 	})
 }
 
@@ -44,4 +60,8 @@ func MatchMemberOf(expected TrustDomain) Matcher {
 		}
 		return nil
 	})
+}
+
+func matchPathPrefix(actual, expected string) bool {
+	return actual == expected || expected == "" || strings.HasPrefix(actual, expected+"/")
 }
--- a/spiffetls/tlsconfig/authorizer_test.go
+++ b/spiffetls/tlsconfig/authorizer_test.go
@@ -1,0 +1,20 @@
+package tlsconfig_test
+
+import (
+	"testing"
+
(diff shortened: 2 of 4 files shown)


matheuscscp:
@amartinezfayo I pushed https://github.com/spiffe/go-spiffe/pull/406/commits/cd621b7b1044ab01a5f2e88536f216ab3d878529 to address copilot's comments

spiffeid/match.go
```diff
@@ -45,3 +61,7 @@ func MatchMemberOf(expected TrustDomain) Matcher {
 		return nil
 	})
 }
+
+func matchPathPrefix(actual, expected string) bool {
+	return actual == expected || expected == "" || strings.HasPrefix(actual, expected+"/")
+}
```
amartinezfayo, line 72: An empty expected path makes this true for any path, so a pathless prefix authorizes the whole trust domain. FromPath accepts an empty path without error, so I think a config-built prefix could reach that state unintentionally. Should we reject a pathless prefix here?

spiffetls/tlsconfig/authorizer_test.go
```diff
@@ -0,0 +1,38 @@
...
+	subpath := spiffeid.RequireFromString("spiffe://example.org/spire/agent/node-a")
+	segmentBoundaryMismatch := spiffeid.RequireFromString("spiffe://example.org/spire/agentish")
+	otherTrustDomain := spiffeid.RequireFromString("spiffe://other.org/spire/agent")
+
+	assert.NoError(t, authorizer(exact, nil))
+	assert.NoError(t, authorizer(subpath, nil))
+	assert.EqualError(t, authorizer(segmentBoundaryMismatch, nil),
+		`unexpected ID "spiffe://example.org/spire/agentish"`)
+	assert.EqualError(t, authorizer(otherTrustDomain, nil),
+		`unexpected ID "spiffe://other.org/spire/agent"`)
```
amartinezfayo: nit: these cases share a shape, and the tests in this package collapse that into a table with a name per case, as TestVerifyPeerCertificate does. Would a single table covering both the prefix-with-path and the pathless variants read better here?

amartinezfayo (commented):
Thank you @matheuscscp!

matheuscscp:
@amartinezfayo Addressed both your comments here: https://github.com/spiffe/go-spiffe/pull/406/commits/d259d6c307d10c190d0e7fba9d325881bdbbc588

spiffeid/match.go
```diff
@@ -22,6 +25,23 @@ func MatchID(expected ID) Matcher {
...
 
+// MatchIDPrefix matches any SPIFFE ID with the given ID prefix. A matching ID
+// must be in the same trust domain and have either the same path as the prefix
+// or a path with the prefix on a segment boundary. The prefix must have a
+// non-empty path or this function panics. Use MatchMemberOf to match any ID in
+// a trust domain.
+func MatchIDPrefix(expected ID) Matcher {
+	if expected.Path() == "" {
+		panic(errPrefixMissingPath)
+	}
```
amartinezfayo, line 37: This does fail closed, though the only other panic in the library is behind the Require functions, and it triggers on a valid SPIFFE ID a caller could have built from config. What do you think about returning a Matcher that always fails with errPrefixMissingPath?

matheuscscp: Fixed here: https://github.com/spiffe/go-spiffe/pull/406/commits/9a042a8ef0e974dfc6504d1f4fb0fd982114fe66

Thanks!

amartinezfayo (approved):
🚀

PR #407, opened 2026-07-24
Enforce audience presence in JWT-SVID validation

bcleenders:
[JWT-SVID §3.2](https://github.com/spiffe/spiffe/blob/main/standards/JWT-SVID.md#32-token-format) requires that tokens carry an `aud` claim and §7.2 requires validators to check it: _"The aud claim MUST be present, containing one or more values. Validators MUST reject tokens without an aud claim set, [...]"_

`parse()` already asserted `sub` and `exp` but not `aud`. It relied on go-jose's `claims.Validate` to enforce audience, but go-jose v4 skips the audience check when `Expected.AnyAudience` is empty:

https://github.com/go-jose/go-jose/blob/8d4e64dd6193f330ff4babb4038c99c56974abff/jwt/validation.go#L92-L104

This means a nil or empty `audience` argument silently disabled all audience validation, and tokens without an `aud` claim were accepted.

Fix:

1. Reject empty `audience` argument at the top of parse(), before any token processing. All internal callers already pass non-empty slices.
2. Reject tokens with no `aud` claim in the same switch that checks `sub` and `exp`, so the check is independent of go-jose behavior.

### DIFF
--- a/svid/jwtsvid/svid_test.go
+++ b/svid/jwtsvid/svid_test.go
@@ -64,8 +64,9 @@ func TestParseAndValidate(t *testing.T) {
 		svid          *jwtsvid.SVID
 	}{
 		{
-			name:   "success",
-			bundle: bundle1,
+			name:     "success",
+			bundle:   bundle1,
+			audience: []string{"audience"},
 			generateToken: func(tb testing.TB) string {
 				claims := jwt.Claims{
 					Subject:  spiffeid.RequireFromPath(trustDomain1, "/host").String(),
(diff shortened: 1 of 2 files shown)


svid/jwtsvid/svid_test.go
```diff
@@ -418,7 +508,8 @@ func TestParseInsecure(t *testing.T) {
 			err: `jwtsvid: token has an invalid subject claim: scheme is missing or invalid`,
 		},
 		{
-			name: "success",
+			name:     "invalid typ",
```
bcleenders, line 511: small drive-by fix; this test was incorrectly named, as you see `generateToken` gets `"invalid"` passed as its fourth argument (`typ`).

PR #409, opened 2026-08-04
Add selector support to broker api

kfox1111:


exp/proto/spiffe/broker/api.proto
```diff
@@ -105,6 +105,38 @@ message KubernetesObjectReference {
...
+// implementation-defined; this specification does not mandate a particular
+// selector vocabulary.
+message Selector {
+    // Required. The type of the selector. MUST NOT be empty and MUST NOT
+    // contain the `:` character.
+    string type = 1;
+
+    // Required. The value of the selector. MUST NOT be empty.
+    string value = 2;
+}
```
sorindumitru, line 119: I think this would be better suited for `spire-api-sdk` instead of `go-spiffe` since it deals with SPIRE concepts, i.e. selectors.

kfox1111: is it possible for it to be defined in a proto in spire-api-sdk and then imported here? does that create a dependency loop?

sorindumitru: It wouldn't be imported here at all. All users of the reference will import it from `spire-api-sdk`.

kfox1111: hmm.. it doesnt need to be referenced in api.proto at all? if not, where would you recommend the proto go in spire-api-sdk? I'm not too familiar with grpc yet.

sorindumitru: https://github.com/spiffe/spire-api-sdk/tree/main/proto/spire/api/types seems like a good place to have this. You can also reuse the existing `Selector` type from there, so you need to specify just the `SelectorReference` type, I'd say in a new file `selectorreference.proto` there.

In the go code where you use you just import the right thing. In the agent code you'd look at the `TypeUrl` field as it's done in the [k8s attestor](https://github.com/spiffe/spire/blob/f380fd78cf7139502b0e8aa05c47236355f441fb/pkg/agent/plugin/workloadattestor/k8s/k8s.go#L623).

kfox1111: Closing in favor of https://github.com/spiffe/spire-api-sdk/pull/101 as per your suggestion.

PR #410, opened 2026-08-04
Allow custom claims on WIT-SVIDs

MarcosDY:
- Store custom claims from WIT-SVIDs into SVID object
- Allow getting WIT-SVID with default spiffeID

### DIFF
--- a/exp/svid/witsvid/source.go
+++ b/exp/svid/witsvid/source.go
@@ -2,8 +2,12 @@ package witsvid
 
 import "github.com/spiffe/go-spiffe/v2/spiffeid"
 
-// Source is a source of WIT-SVIDs keyed by SPIFFE ID.
+// Source is a source of WIT-SVIDs.
 type Source interface {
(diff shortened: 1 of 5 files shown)


exp/svid/witsvid/source.go
```diff
@@ -2,8 +2,12 @@ package witsvid
 
 import "github.com/spiffe/go-spiffe/v2/spiffeid"
 
-// Source is a source of WIT-SVIDs keyed by SPIFFE ID.
+// Source is a source of WIT-SVIDs.
 type Source interface {
+	// GetWITSVID returns the default WIT-SVID from the source. It is used by
+	// code that presents a credential and therefore has no ID to look up.
```
sorindumitru: I'm not sure I understand the second sentence, there's no credential to present either way. The first sentence seems good enough so you could leave it at that.

MarcosDY: yeap, it is why i added it because i need into another pr and in caes I dont know the spiffeID lets me remvoe that.

PR #412, opened 2026-08-15
Add WPT authentication for WIT-SVIDs

MarcosDY:
Adds exp/witwpt, the transport-neutral implementation of the Workload Proof Token protocol, and exp/withttp, its net/http integration. 
A client mints a short-lived proof signed with the WIT-SVID's cnf key; a verifier validates the credential against its trust bundle before checking that proof.

Also adds a Claims field to witsvid.SVID, populated from data parse
already had in hand.

Depends of https://github.com/spiffe/go-spiffe/pull/410

### DIFF
--- a/exp/svid/witsvid/source.go
+++ b/exp/svid/witsvid/source.go
@@ -2,8 +2,12 @@ package witsvid
 
 import "github.com/spiffe/go-spiffe/v2/spiffeid"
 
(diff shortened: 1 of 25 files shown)


MarcosDY:
I'm splitting this into multiple PRs to avoid a 3k lines PR

PR #413, opened 2026-08-16
Create lite version of go-spiffe

kfox1111:


amartinezfayo:
Thanks for taking this on @kfox1111! I think this aligns well with the multi-module approach discussed in https://github.com/spiffe/spiffe/issues/414#issuecomment-5058648605. The lite module's importable packages depend only on go-jose, and the v2 module depends on it while re-exporting every symbol at the original import paths. I verified that each re-exported package exposes the same exported symbols as v2.8.1, so existing v2 importers need no changes now or later, and adopting lite stays an opt-in import path rewrite for consumers that want to shed gRPC.

The one structural thing I think we should settle before this leaves draft is the root go.mod. While the lite require points at v0.0.0, a consumer pinning go-spiffe/v2 at a commit or branch cannot resolve the lite dependency, because replace directives in a dependency's go.mod are ignored and no lite tag exists yet. The replace does still apply to the GOWORK=off CI build, so that job does not yet build the way a consumer would. I think the steady state we want is the require always pointing at a real lite tag, which is how opentelemetry-go and google-cloud-go pin their nested modules. Since this PR already adds go.work for in-repo development, I would also drop the replace once a tag exists (google-cloud-go does without replaces entirely), which turns the GOWORK=off build into a true consumer check. To get there without ever merging an unresolvable go.mod, I suggest we split this into two PRs. The first adds the lite module as a copy without touching v2, so nothing imports it yet and main stays consumable at every commit, and we tag lite from that. The second converts the v2 packages to aliases, pins the require to the tag, and deletes the duplicated code.
What do you think?

kfox1111:
I've been thinking about that... There is one big drawback to that approach... It makes it look like I copied all the code rather then just renamed it.... And then in the second pr, it deletes all the history I think. Git blame on a file no longer links lines to commits, it all belongs to the one commit.

What if... we juggled a little bit. We review the pr as is as if it was going to merge to `main`, and make sure we are happy with it. We then merge it upstream as branch 'lite' rather then merge to 'main'. We then have a usable referancable lite branch. Nothing points to it yet.

We then make a pr against `lite` to fix the reference. It should be fine as it is an existing branch at this point.

Then we merge lite to main. This should be a pure fast forward merge, so should be safe.

Now both branches point at the same commit with valid references to `lite`.

PR #415, opened 2026-08-22
Add exp/witwpt with WPT minting and policy helpers

MarcosDY:
Add exp/witwpt, the WPT protocol core: minting and policy

Introduces the transport-neutral core of WPT (workload proof token)   authentication for WIT-SVIDs, covering the presenting side and the   policy helpers a verifier composes with.

### DIFF
--- a/exp/witwpt/mint.go
+++ b/exp/witwpt/mint.go
@@ -1,0 +1,160 @@
+// Package witwpt implements the Workload Proof Token (WPT) proof-of-possession
+// protocol for WIT-SVIDs, as specified by draft-ietf-wimse-wpt-01. A workload
+// mints a short-lived WPT signed with the private key matching the WIT-SVID's
+// cnf.jwk claim, binding the proof to the credential (wth) and to the request
+// target (aud).
+//
+// This package is transport neutral. See exp/withttp for the net/http
+// integration.
+package witwpt
+
+import (
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/go-jose/go-jose/v4"
+	"github.com/go-jose/go-jose/v4/jwt"
+	"github.com/spiffe/go-spiffe/v2/exp/svid/witsvid"
+)
+
+const (
+	// DefaultProofLifetime is how far ahead a minted proof's exp is set. The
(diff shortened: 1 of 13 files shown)


exp/witwpt/mint.go
```diff
@@ -0,0 +1,160 @@
...
+func (fn mintOption) configureMint(c *mintConfig) { fn(c) }
+
+// WithProofLifetime sets how far ahead the minted proof's exp is placed,
+// overriding DefaultProofLifetime. Raising it past the verifier's maximum
+// lifetime causes the proof to be rejected.
+func WithProofLifetime(lifetime time.Duration) MintOption {
+	return mintOption(func(c *mintConfig) {
+		c.lifetime = lifetime
+	})
+}
```
sorindumitru, line 66: Having the option to include extra claims would be good to have from the beginning.

exp/witwpt/mint.go
```diff
@@ -0,0 +1,160 @@
...
+		return "", wrapErr(fmt.Errorf("unable to build proof signer: %w", err))
+	}
+
+	now := time.Now()
+	proof, err := jwt.Signed(signer).Claims(map[string]any{
+		"aud": audience,
+		"exp": jwt.NewNumericDate(now.Add(config.lifetime)),
+		"iat": jwt.NewNumericDate(now),
+		"jti": jti,
+		"wth": WITThumbprint(witToken),
```
sorindumitru, line 119: I don't know if we want to do this now, but we can could store the thumbprint in the SVID type, to avoid recomputing it for each issues WPT.

exp/witwpt/mint.go
MarcosDY: I agree with have this from the start, but the verifier in the follow-up PR decodes only aud/exp/iat/jti/wth and drops the rest, so WithClaims here would write claims nothing can read yet.
I prefer to include that code on the next PR that I already have but I need to wait for stacked PRs. (we can release once all is in place)

exp/witwpt/mint.go
MarcosDY: The cost is not that bad, I prefer to no add "for now" until we see it is really affecting, (we can introduce once all is in place and we can test end to end to verify if it is really a bottleneck, but I dont think it is going to be a big one.
I'm worry about caching it is not worth the token/thumbprint staleness risk.

PR #422, opened 2026-09-16
feat: allow retrieving PeerID before first read

oliverpool:
By calling `conn.Handshake` manually, we can make sure that the handshake has been completed.

It is currently quite akward to first read some data before being able to identify the peer.

It shouldn't have any performance impact, since `conn.Handshake` is already called on every `conn.Read` inside the Go TLS library: https://github.com/golang/go/blob/ffd88e127c37923e6fe2e43f7fde7e5d1501094c/src/crypto/tls/conn.go#L1481

The adjusted test fails without this patch.


spiffetls/listen.go
```diff
@@ -143,9 +143,13 @@ type serverConn struct {
 	*tls.Conn
 }
 
-// PeerID returns the peer SPIFFE ID on the connection. The handshake must have
-// been completed. Note that in Go's TLS stack, the TLS 1.3 handshake may not
-// complete until the first read from the connection.
+// PeerID returns the peer SPIFFE ID on the connection.
```
amartinezfayo, line 146: I think PeerID can now block on network I/O when the handshake has not run yet, bounded only by the connection deadlines. Would it make sense to say so in the comment so callers know to set a deadline first?

spiffetls/listen.go
```diff
@@ -143,9 +143,13 @@ type serverConn struct {
...
 func (c *serverConn) PeerID() (spiffeid.ID, error) {
+	// The TLS handshake may not be completed until the first read,
+	// but is required to populate the PeerCertificates.
+	// Trigger it manually (like *tls.Conn.Read).
+	if err := c.Handshake(); err != nil {
+		return spiffeid.ID{}, wrapSpiffetlsErr(fmt.Errorf("tls.Handshake: %w", err))
```
amartinezfayo, line 152: nit: The other messages in this package read as prose, like `unable to dial` in DialWithMode and `unable to close TLS connection` in Close. Maybe we could use something like `unable to complete TLS handshake` here for consistency?

spiffetls/peerid.go
```diff
@@ -15,15 +15,17 @@ type PeerIDGetter interface {
 }
 
 // PeerIDFromConn returns the peer ID from a server or client peer connection.
```
amartinezfayo, line 17: nit: clientConn.PeerID still carries the handshake note this PR removes from the server side, even though tls.Dial completes the handshake before returning. Should we update that comment too, or call Handshake there as well for symmetry?

spiffetls/listen.go
oliverpool: Note added in dba6d94dd7bd389745af5bb0332acb391f2bd8d7

spiffetls/listen.go
oliverpool: Adjusted in dba6d94dd7bd389745af5bb0332acb391f2bd8d7

spiffetls/peerid.go
oliverpool: Good catch, I adjusted the client accordingly: c44fa00912cb682f3270c3751bbcd49fa1d03543

oliverpool:
@amartinezfayo thanks for the review, I adjusted the comments and the code (had to force-push, because I forgot to signoff: the commit IDs don't match anymore, sorry...)

amartinezfayo (approved):
Thank you @oliverpool!

oliverpool:
There is something wrong with the merge train or the CI 🤔

